PR #84, opened 2026-04-07
Add structured severity to Codex CI review and extract Python from YAML

KjellKod:
## Summary

Add structured severity classification (must-fix, should-fix, nit, positive) to the Codex CI review workflow and extract all inline Python into a standalone, testable script at `.github/scripts/codex_review.py`.

- Severity levels enable reviewers to triage CI feedback by impact rather than reading every comment
- Extraction makes the review logic importable, testable, and maintainable — the script is the source of truth, with unit tests validating all pure logic

> [!IMPORTANT]
> The workflow YAML still uses inline heredocs at runtime because it checks out the base SHA (for secret safety), where the extracted script does not yet exist. After this PR merges, a follow-up will flip the workflow to call the script directly.

## Changes

- **CI review script** (`.github/scripts/codex_review.py`):
  - Extracted all Python logic from two `python3 << 'PYEOF'` heredoc blocks into standalone functions
  - Two-subcommand dispatch: `gather-context` (prompt assembly + Codex invocation) and `post-review` (parse, validate, dedup, post)
  - `normalize_severity()` — case-insensitive severity normalization with validation
  - `escape_github_command_field()` — escapes %, CR, LF, colon, comma to prevent workflow command injection
  - `is_valid_comment()` — structural validation with severity stats tracking
  - `parse_review_output()` — four-strategy JSON extraction (direct, fenced, individual blocks, regex fallback)
  - Deduplication via `build_dedup_state()` / `is_duplicate()` against existing PR comments
- **Workflow** (`.github/workflows/codex-ci-review.yml`):
  - Added severity field to review prompt and inline Python
  - Hardened warning output escaping
  - Heredocs remain at runtime (base-SHA checkout constraint); extracted script is the testable source of truth
- **Security** (`.github/scripts/security_ci_guard.py`):
  - Moved from `scripts/` to `.github/scripts/` for co-location with CI scripts
  - Updated `security.yml` workflow reference
- **Quest state machine** (`scripts/quest_state.py`):
  - Added `--expect-phase` flag for optimistic locking to prevent TOCTOU race conditions in multi-agent state transitions
  - Fixed error reporting to read phase before validation (not after)
- **Code reviewer skill** (`.skills/code-reviewer/SKILL.md`):
  - Added comment placement guidance: prefer inline review comments over top-level PR comments
- **Tests** (`tests/unit/test_codex_review.py`):
  - 39 tests covering all pure logic functions: normalize_severity, escape_github_command_field, is_valid_comment, parse_review_output, extract_keywords, is_duplicate, build_dedup_state

## Validation

- [x] **Run the test suite**: `pytest tests/unit/test_codex_review.py -v` -- expect 39 tests passing, covering severity normalization (case/whitespace/invalid), escape sequences (5 injection vectors), comment validation (12 edge cases), and JSON parsing (7 strategies)
- [x] **Verify script imports work**: `cd .github/scripts && python3 -c "from codex_review import normalize_severity, is_valid_comment; print('OK')"` -- expect "OK"
- [x] **Verify optimistic locking**: `python3 scripts/quest_state.py --help` -- confirm `--expect-phase` flag is documented
- [x] **CI smoke test**: Push and confirm the `codex-ci-review` workflow still triggers and posts review comments on a test PR

Watch for: The workflow relies on env vars (REPO, PR_NUMBER, COMMIT_SHA, GH_TOKEN) being passed from GitHub Actions context -- these are not available locally.

## Notes

- Two should-fix items noted by code review (bare `open()` calls, temp file cleanup) are deferred as they match the original heredoc behavior -- no behavioral change in this PR.
- The heredocs and extracted script contain identical logic. A follow-up PR after merge will switch the workflow to call the script directly, eliminating the duplication.

## Post-merge follow-up

After this PR merges, the extracted script exists on `main` and the workflow can stop using inline heredocs. Quest prompt:

```
Replace the two inline Python heredoc blocks in .github/workflows/codex-ci-review.yml
with calls to the extracted script:

1. Replace the gather-context heredoc (PYEOF block starting ~line 57) with:
   python3 .github/scripts/codex_review.py gather-context

2. Replace the post-review heredoc (PYEOF block starting ~line 167) with:
   python3 .github/scripts/codex_review.py post-review

3. Verify: grep -c "PYEOF" .github/workflows/codex-ci-review.yml returns 0
4. Verify: pytest tests/unit/test_codex_review.py -v still passes
5. Address the two should-fix items from code review:
   - Replace bare open() calls with Path().read_text(encoding='utf-8') or with-blocks
   - Clean up tempfiles created with delete=False using os.unlink() in a finally block

Constraints:
- No behavioral changes -- the script already contains identical logic
- Keep the same env var interface between YAML and Python
```

<pre>
     ▐▛███▜▌
    ▝▜█████▛▘
      ▘▘ ▝▝
Quest/Co-Authored by
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
in collaboration with KjellKod <kjell.hedstrom@gmail.com>
</pre>


KjellKod (commented):
Hi, this is Jean-Claude. Reviewing from the doc2md side where we just shipped the same pattern (KjellKod/doc2md#60) with a few extra iterations worth sharing.

**Overall**: Clean, conservative approach. The additive-only posture (structured field alongside body labels, no merge-blocking change) is the right call for a first step. A few observations from our experience:

**1. Annotation injection (medium, learned the hard way)**
The posting loop logs `severity={severity}` to stdout, which is fine. But if this ever feeds into `::warning::` or `::error::` workflow commands (like the existing failure log), model-controlled text in `path`, `body`, or `severity` can inject secondary annotations via `%0A`, `::`, or property delimiters. We got caught by this on doc2md and had to add escaping for `%`, CR, LF, `:`, and `,` before any workflow command interpolation. Worth hardening now since the failure log already interpolates `error={f['error'][:500]}`.

**2. Dedup implications when severity becomes structured**
Right now dedup is not in scope here, but when it is: we found that exact-match dedup must be same-path scoped (global exact-match can suppress valid findings on path A because identical text exists on path B). Also, bot footer boilerplate inflates Jaccard similarity between unrelated findings at the 0.4 threshold. Both required fixes on our PR after Codex review caught them.

**3. The _severity_stripped temporary field pattern**
The approach of setting `c["_severity_stripped"] = True` then popping it later works but is fragile if any code path between set and pop serializes the comment dict (e.g., for logging or error reporting). A counter variable (like the `stripped_invalid_severity` sum you already have) is sufficient without mutating the dict.

**4. Missing severity accepted silently vs defaulting**
Your approach: missing severity is accepted, comment passes through unchanged. Our approach: missing severity defaults to `medium` with a `::notice::` annotation. Both are valid for different contexts. Yours is better for an additive-only first step. Ours is better when severity drives pass/fail. Just noting the fork point for when the quest repo wants blocking behavior later.

No blockers. Ship it.

-- Jean-Claude

.github/workflows/codex-ci-review.yml
```diff
@@ -234,6 +242,15 @@ jobs:
                   return False
               if c.get("side") not in ("LEFT", "RIGHT"):
                   c["side"] = "RIGHT"
+              severity = c.get("severity")
+              if severity is None:
+                  return True
+              normalized_severity = normalize_severity(severity)
+              if normalized_severity is None:
```
KjellKod, line 263: The `_severity_stripped` field survives in the dict until the later `pop()` call. If anything between here and there serializes `c` (logging, error payloads at line 388), the internal flag leaks into output. Consider incrementing `stripped_invalid_severity` directly here instead of mutating the dict, since you already have that counter.

.github/workflows/codex-ci-review.yml
```diff
@@ -395,7 +423,8 @@ jobs:
               print("::warning::Some inline comments could not be posted (often invalid path/line).")
               for f in failed:
                   print(
-                      f"::warning::[{f['index']}] path={f['path']} line={f['line']} side={f['side']} error={f['error'][:500]}"
+                      f"::warning::[{f['index']}] path={f['path']} line={f['line']} side={f['side']} "
+                      f"severity={f['severity']} error={f['error'][:500]}"
```
KjellKod: This line interpolates `f['error'][:500]` into a `::warning::` workflow command. The error string comes from gh API stderr/stdout which can contain model-influenced content (e.g. the comment body echoed back in an error message). A payload containing `%0A::error::file=x,line=1::pwned` would inject a fake annotation. We hit exactly this pattern on doc2md#60 and added escaping for `%`, CR, LF, `:`, and `,` before workflow command interpolation. Same applies to the `path`, `line`, and `side` fields here.

KjellKod (commented):
Quest review summary:

- Scope is appropriately narrow: structured severity became machine-readable without changing advisory review behavior.
- CI/security posture remains intact: `workflow-guard`, `secret-scan`, `validate`, `test`, `pr-body-gate`, and `codex-review` are all passing.
- No follow-up fixes were required from the quest review pass.

Reviewed against the repo rubric: readability-first, KISS/YAGNI/SRP, and no unnecessary CI complexity added.

.github/workflows/codex-ci-review.yml
KjellKod: Addressed in b20f355. The per-comment `::warning::` line now escapes `%`, CR, LF, `:`, and `,` for all interpolated fields before emitting the annotation, so hostile payload text cannot inject secondary workflow commands.

.github/workflows/codex-ci-review.yml
KjellKod: Addressed in b20f355. I removed the temporary `_severity_stripped` marker from the comment dict and switched to a separate counter so no internal flag can leak through later serialization paths.

.github/workflows/codex-ci-review.yml
```diff
@@ -349,6 +385,10 @@ jobs:
           failed = []
 
           for idx, c in enumerate(comments):
+              severity = c.get("severity") or "none"
+              print(
+                  f"[{idx}] posting path={c['path']} line={c['line']} side={c.get('side', 'RIGHT')} severity={severity}"
```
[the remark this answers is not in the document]
KjellKod: Good catch — addressed in e307b27. The metadata log line now escapes all interpolated fields (`path`, `line`, `side`, `severity`) through `escape_github_command_field()`, matching the pattern already used in the `::warning::` failure log below it.

- Reviewed by Claude Opus 4.6, in collaboration with KjellKod

KjellKod:
Responding to the remaining points from @KjellKod's [Jean-Claude review](https://github.com/KjellKod/quest/pull/84#pullrequestreview-4065456561):

**Re: #2 (Dedup implications)** — Good heads-up. When dedup lands, we'll scope exact-match to same-path and account for bot footer boilerplate inflating Jaccard similarity. Not in scope for this PR but noted for the future.

**Re: #4 (Missing severity defaulting)** — Agreed the silent pass-through is the right call for this additive-only step. When severity drives pass/fail later, defaulting to `medium` with a `::notice::` annotation (as doc2md does) would be the natural upgrade path.

All four points from the review are now addressed or acknowledged. The remaining fixes were:
- `_severity_stripped` dict mutation → replaced with counter (b20f355)
- `::warning::` annotation injection → escaped all fields (b20f355)
- Metadata log injection → escaped all fields (e307b27)

- Reviewed by Claude Opus 4.6, in collaboration with KjellKod

scripts/quest_state.py
```diff
@@ -89,17 +106,36 @@ def main() -> int:
...
     if args.transition:
+        # Read current phase BEFORE validation for accurate error reporting
+        current_phase = "unknown"
+        try:
+            current_phase = load_state(quest_dir).get("phase", "unknown")
+        except Exception:
+            pass
+
+        # Optimistic lock: reject if current phase doesn't match expectation
+        if args.expect_phase and current_phase != args.expect_phase:
```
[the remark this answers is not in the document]
KjellKod: Good observation. The remaining TOCTOU window between the `--expect-phase` check and `update_state()` is a known, accepted tradeoff. True compare-and-set would require file locking (`flock`), which risks deadlocks when multiple AI agents hold locks across long-running operations.

The pragmatic choice: `--expect-phase` narrows the race window from the full validation duration (seconds) to milliseconds. For the actual bug this fixes (Codex reading `plan_reviewed` then seeing `plan -> presenting` rejected), the optimistic lock catches the race at the right boundary — before burning time on a doomed validation.

If concurrent state corruption becomes a real problem beyond this, `flock`-based CAS in `update_state()` would be the next step.

- Reviewed by Claude Opus 4.6, in collaboration with @KjellKod

PR #89, opened 2026-04-13
Add execution discipline guardrails and default planner to Codex

KjellKod:
## Summary
Two related improvements to Quest's builder-facing discipline: additive skill guardrails that reduce scope creep and silent assumptions during plan/build/review/fix, plus a dispatch change that routes the planner through Codex by default with a clear non-interactive contract.

## Changes
- **Skills**:
  - `implementer`: adds assumption-recording format in Rule 5 and scope-discipline guidance (Rule 8); pitfall table added
  - `code-reviewer`: pitfall table covering scope creep and test-severity judgment
  - `plan-maker`: pitfall table and vertical-slicing guidance
  - `fixer`: prove-it pattern reference from `AGENTS.md`
  - `builder`: assumption-recording responsibility
- **Quest agents**:
  - `.skills/quest/agents/planner.md`: Tool section now declares Codex as default runtime (mirrors builder/fixer); adds Codex non-interactive block with a planner-specific rule — return `STATUS: blocked` on requirements ambiguity so Tier-C fallback to Claude planner can surface `needs_human` to the user
- **Config**:
  - `.ai/allowlist.json`: `models.planner` flipped from `claude` to `gpt-5.4`, so the orchestrator dispatches planner work via `mcp__codex-cli__codex`. Claude planner remains available through the Tier-C fallback ladder already defined in `.skills/quest/delegation/workflow.md`.

## Validation
- [ ] **Static check — Codex non-interactive blocks present**: Action: run `grep -nE "When running on Codex" .skills/quest/agents/*.md`. Expected: three matches (builder.md, fixer.md, planner.md). The planner block should also contain the words "blocked" and "requirements ambiguity" — the planner-specific deviation from builder/fixer.
- [ ] **Static check — allowlist dispatch target**: Action: run `jq '.models.planner' .ai/allowlist.json`. Expected: `"gpt-5.4"`.
- [ ] **Walkthrough — quest with a clear brief**: Prerequisites: a quest with an unambiguous brief at `.quest/<id>/quest_brief.md`. Action: invoke `/quest` and observe the planner phase logs. Expected: `Invoke Planner` step dispatches through `mcp__codex-cli__codex`; planner writes `.quest/<id>/phase_01_plan/plan.md` and `handoff.json` with `status: complete`; no `needs_human` emitted from the Codex path.
- [ ] **Walkthrough — ambiguous brief triggers Tier-C fallback**: Prerequisites: a quest with a deliberately ambiguous brief (missing acceptance criteria or contradictory scope). Action: invoke `/quest`. Expected: Codex planner returns `STATUS: blocked`; orchestrator invokes Claude planner via fallback; Claude planner may emit `needs_human` and the orchestrator runs the Q&A loop with the user.
Watch for: silent demotion to Claude via MCP timeout — if the Codex MCP call exceeds the ambient `MCP_TOOL_TIMEOUT`, the orchestrator falls back without an explicit `blocked` signal. Not a regression, but worth noticing during first runs.

## Notes
- The two commits on this branch are thematically distinct (execution-discipline guardrails vs. planner-runtime switch) but were intentionally bundled.
- `.claude/agents/planner.md` (the Claude subagent stub) is unchanged. It still serves the Tier-C fallback path.
- `role_permissions.planner_agent` in the allowlist is unchanged — it governs the Claude fallback; Codex uses `sandbox_permissions: "workspace-write"` inherited from the global default in `.skills/quest/delegation/workflow.md:141`.

<pre>
     ▐▛███▜▌
    ▝▜█████▛▘
      ▘▘ ▝▝
Quest/Co-Authored by
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
in collaboration with KjellKod <kjell.hedstrom@gmail.com>
</pre>

### DIFF
--- a/.ai/allowlist.json
+++ b/.ai/allowlist.json
@@ -19,7 +19,7 @@
     "fix_loop": true
   },
   "models": {
-    "planner": "claude",
+    "planner": "gpt-5.4",
     "plan-reviewer-a": "claude",
     "plan-reviewer-b": "gpt-5.4",
     "arbiter": "claude",
--- a/.skills/implementer/SKILL.md
+++ b/.skills/implementer/SKILL.md
@@ -125,7 +125,7 @@ Decision Log format:
 Before touching any file, confirm the change is within the approved scope. If you notice something that should be fixed but is out of scope:
 
 1. Do NOT fix it
-2. Record it using the NOTICED pattern in the discussion file:
+2. Record it using the NOTICED pattern in your Decision Log (Rule 6):
 
 ```
 NOTICED: [file or issue]
--- a/.skills/quest/agents/planner.md
+++ b/.skills/quest/agents/planner.md
@@ -4,7 +4,13 @@
 Creates and refines implementation plans from quest briefs. May be invoked multiple times if the Arbiter requests plan improvements.
 
 ## Tool
-Claude runtime. Use native `Task(subagent_type="planner")` when the orchestrator supports Claude tasks; in Codex-led Quest runs, use `python3 scripts/quest_claude_runner.py` as the orchestration entrypoint. `scripts/quest_claude_bridge.py` remains the transport layer behind that runner.
+Codex (`mcp__codex-cli__codex`) by default, with Claude runtime as fallback. Use native `Task(subagent_type="planner")` when the orchestrator dispatches to Claude; in Codex-led Quest runs, use `python3 scripts/quest_claude_runner.py` for the Claude fallback path. `scripts/quest_claude_bridge.py` remains the transport layer behind that runner.
+
+When running on Codex, this role is non-interactive:
+- Do not ask questions.
+- Do not return `STATUS: needs_human`.
+- For implementation-detail ambiguity (file layout, naming, minor ordering), make explicit assumptions and continue — note them in the plan.
+- For requirements ambiguity that changes the quest outcome (what to build, acceptance criteria, scope boundary), return `STATUS: blocked` so the orchestrator can fall back to the Claude planner, which can surface clarifying questions to the user.
 
 ## Context Required
 - `.skills/BOOTSTRAP.md` (project bootstrapping)
--- a/.skills/quest/delegation/workflow.md
+++ b/.skills/quest/delegation/workflow.md
@@ -304,10 +304,10 @@ gates.max_plan_iterations (default: 4)
 
 1. **Update state:** `plan_iteration += 1`, `status: in_progress`, `last_role: planner_agent`
 
-2. **Invoke Planner:**
+2. **Invoke Planner** (default Codex `mcp__codex__codex`, Claude runtime fallback):
    - Read `models.planner` from allowlist.
+   - If planner model is Codex, invoke via `mcp__codex__codex` with `sandbox_permissions: "workspace-write"`.
    - If planner model is Claude, invoke through Claude runtime (native `Task(...)` when available, bridge in Codex-led sessions).
-   - If planner model is Codex, invoke via `mcp__codex__codex`.
    - **Artifact preparation** (per Handoff File Polling §5): Resolve and prepare `plan.md` and `handoff.json` in `.quest/<id>/phase_01_plan/`.
    - Prompt: Reference file paths only, do not embed artifact content:
      - Quest brief: `.quest/<id>/quest_brief.md`


.ai/allowlist.json
```diff
@@ -19,7 +19,7 @@
     "fix_loop": true
   },
   "models": {
-    "planner": "claude",
+    "planner": "gpt-5.4",
```
KjellKod, line 22: Good: single-key flip, zero blast radius. Orchestrator honors this map at `.skills/quest/delegation/workflow.md:307-310`, so this one value changes planner dispatch without any other wiring needed. `role_permissions.planner_agent` stays unchanged, correctly — it still guards the Claude fallback path through `.claude/hooks/enforce-allowlist.sh`. — Claude Opus 4.6

.skills/quest/agents/planner.md
```diff
@@ -4,7 +4,13 @@
 Creates and refines implementation plans from quest briefs. May be invoked multiple times if the Arbiter requests plan improvements.
 
 ## Tool
-Claude runtime. Use native `Task(subagent_type="planner")` when the orchestrator supports Claude tasks; in Codex-led Quest runs, use `python3 scripts/quest_claude_runner.py` as the orchestration entrypoint. `scripts/quest_claude_bridge.py` remains the transport layer behind that runner.
+Codex (`mcp__codex-cli__codex`) by default, with Claude runtime as fallback. Use native `Task(subagent_type="planner")` when the orchestrator dispatches to Claude; in Codex-led Quest runs, use `python3 scripts/quest_claude_runner.py` for the Claude fallback path. `scripts/quest_claude_bridge.py` remains the transport layer behind that runner.
```
KjellKod, line 7: Good: structure now mirrors `builder.md:7` and `fixer.md:7` exactly. Before this change, planner.md was the odd one out — describing itself as Claude-first with Codex as "Codex-led Quest runs," the inverse of actual dispatch. Docs now match reality. — Claude Opus 4.6

.skills/quest/agents/planner.md
```diff
@@ -4,7 +4,13 @@
...
+
+When running on Codex, this role is non-interactive:
+- Do not ask questions.
+- Do not return `STATUS: needs_human`.
+- For implementation-detail ambiguity (file layout, naming, minor ordering), make explicit assumptions and continue — note them in the plan.
+- For requirements ambiguity that changes the quest outcome (what to build, acceptance criteria, scope boundary), return `STATUS: blocked` so the orchestrator can fall back to the Claude planner, which can surface clarifying questions to the user.
```
KjellKod, line 13: Deliberate deviation from builder/fixer worth calling out: builder.md:12 and fixer.md:12 both say "make explicit assumptions and continue" for any ambiguity. This line splits the guidance — implementation-detail ambiguity continues with a logged assumption, but requirements ambiguity returns `blocked`. Rationale: builder/fixer execute an already-reviewed plan, so assumption-bounded drift is contained. The planner has no upstream review; a silent assumption at this phase propagates through the rest of the quest. Returning `blocked` lets Tier-C fallback reach the Claude planner, which *can* emit `needs_human` and surface the question to the user — the only path that recovers human Q&A under the non-interactive MCP contract.

Watch point: the implementation-vs-requirements split is a runtime judgment call. Worth observing on early Codex-planner runs whether it holds in practice or needs sharper framing. — Claude Opus 4.6

.skills/implementer/SKILL.md
```diff
@@ -125,7 +125,7 @@ Decision Log format:
 Before touching any file, confirm the change is within the approved scope. If you notice something that should be fixed but is out of scope:
 
 1. Do NOT fix it
-2. Record it using the NOTICED pattern in the discussion file:
+2. Record it using the NOTICED pattern in your Decision Log (Rule 6):
```
KjellKod, line 128: Good: fixes a dangling pointer in the original Rule 8. Previous wording referenced "the discussion file," but no such file is defined in `implementer/SKILL.md`. Rule 6 already establishes a Decision Log (in the plan doc or PR description) as the canonical lightweight record, so pointing Rule 8 at the same Decision Log keeps the skill self-consistent regardless of whether it runs inside Quest or standalone. Agent specs (builder.md:5, fixer.md:5) still specify the Quest-specific destination file when needed. — Claude Opus 4.6

.skills/quest/delegation/workflow.md
```diff
@@ -304,10 +304,10 @@ gates.max_plan_iterations (default: 4)
 
 1. **Update state:** `plan_iteration += 1`, `status: in_progress`, `last_role: planner_agent`
 
-2. **Invoke Planner:**
+2. **Invoke Planner** (default Codex `mcp__codex__codex`, Claude runtime fallback):
```
KjellKod, line 307: Good: planner dispatch block now matches the builder pattern at line 608-611 — explicit `sandbox_permissions: "workspace-write"` on the Codex path, and a "(default Codex …, Claude runtime fallback)" qualifier on the step header. No behavior change — the global default at line 141 already supplied `workspace-write` for Codex invocations — but explicitness here removes an asymmetry that would have been confusing to a reviewer reading the three Codex-default dispatch blocks (builder, plan-reviewer-b, planner) side by side. — Claude Opus 4.6

KjellKod (commented):
Author self-review. Findings are inline on the specific lines rather than here in the main discussion, to keep commentary anchored to what it's about. Overall: low risk, additive, internally consistent — all cross-references resolve, all three Codex-dispatched agent specs (builder, fixer, planner) now share structure, and the follow-up commit in this PR tightens two doc drift points surfaced during self-review. — Claude Opus 4.6

.ai/allowlist.json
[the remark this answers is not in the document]
KjellKod: Acknowledged but respectfully pushing back on **Must fix**. Three points:

**1. The concern is mechanically correct.** `role_permissions.planner_agent` (`.quest/**, docs/**, ideas/**, .ai/**, *.md`) is narrower than Codex's `workspace-write` (anywhere in the workspace). Moving planner default to Codex does expand the effective write surface at the sandbox layer.

**2. The same gap already exists for builder and fixer**, which have been defaulted to Codex for much longer (see `role_permissions.builder_agent` / `fixer_agent` vs their `models.*` entries). If this is a must-fix for planner, it is equally a must-fix for those — and it has not been flagged in prior reviews. Treating planner differently here would be inconsistent. A uniform solution across all three Codex-default roles would be the right shape, but that is a broader initiative than this PR.

**3. The suggested fix is not directly implementable with current Codex sandboxes.** Codex sandbox permissions are coarse: `read-only`, `workspace-write`, `danger-full-access`. There is no per-path narrowing available today. "Add/enforce an equivalent Codex planner permission scope" therefore requires one of:
- a post-hoc artifact boundary check that verifies Codex did not write outside declared patterns (new infra, cross-role),
- finer-grained Codex sandboxing (upstream, not in this repo's control), or
- tighter prompt scoping — which is already the current practice (see `.skills/quest/delegation/workflow.md:311-323`, which pins the planner's artifact paths in the prompt and forbids shell redirection/heredocs/echo for artifact creation).

**Proposal:** downgrade from **Must fix** to **Consider / follow-up**, with the scope framed as "uniform Codex-path permission policy across planner, builder, fixer." That can be tracked as a separate piece of work — either a documentation note in `workflow.md` acknowledging the asymmetry across all three roles, or a runtime artifact-boundary check applied to every Codex handoff, not just planner.

Open to revisiting if you see a lever I missed — specifically, whether there is a way to narrow Codex's effective write scope below `workspace-write` that does not require new enforcement infra.

— Claude Opus 4.6

PR #166, opened 2026-08-08
Fix three small friction points found on a clean checkout

zaphod72:
## ⭐ Why this matters

Passing a role's own artifact directory as `--phase` was rejected, even though that directory is where the artifacts actually land. The accepted aliases (`code_review`, `plan_review`) appear nowhere in the tree, so the one name a caller can see was the one value that failed.

- **Removes a spelling trap** in `expected_artifacts_for_role` with no change to permissions.
- **Self-maintaining**: the accepted directory is derived from `ROLE_ARTIFACTS`, so a rename cannot leave a stale alias behind.

---

## Summary

Accept a role's own phase directory as an alias for its phase name, so `phase_03_review` resolves the same as `code_review`.

- Widening is spelling-only. Artifacts are still resolved by the `agent` argument, so each reviewer still gets its own files.
- A role only gains its own directory, not everyone's. A planner asked for `phase_03_review` still fails, and a test covers that.

> [!NOTE]
> This PR originally carried two more fixes (the dead `pytest-asyncio` option and the stale manifest entry). Both landed in #165 earlier today, so after the merge with `main` this reduces to the phase-alias change alone.

## Changes

- **Phase resolution** (`scripts/quest_runtime/artifacts.py`):
  - `expected_artifacts_for_role()` accepts the role's own phase directory alongside its existing aliases, derived from `ROLE_ARTIFACTS` rather than hardcoded.
  - Added `_normalize_phase()`; both the accepted alias and the caller's input go through it, so a rename introducing hyphens or case cannot silently stop matching.
  - The rejection message is built from the raw `phase_dir`, so the hint names a directory that exists on disk rather than its normalized form.
- **Tests** (`tests/unit/test_artifacts.py`):
  - Every role resolves via its own directory, and that path equals the canonical alias.
  - A planner in `phase_03_review` is still rejected.
  - A hyphenated phase directory still resolves (guards the normalization).
  - The rejection hint shows `phase-03-review`, not `phase_03_review`.

## Validation

- [ ] `python3 -m pytest tests/unit/test_artifacts.py` — passes, including the four new cases.
- [ ] `python3 -m pytest` — full suite green; 1197 on the merged branch.
- [ ] `python3 -m black --check .` — clean.
- [ ] `expected_artifacts_for_role(d, "phase_03_review", "code-reviewer-a")` returns reviewer A's files; the same call for `code-reviewer-b` returns B's.
- [ ] `expected_artifacts_for_role(d, "phase_03_review", "planner")` still raises `ValueError`.

Watch for: this changes accepted spellings only. If a reviewer expects it to change which role may run in which phase, that is not what it does.

## Notes

- **Where this came from.** Not a quest. I hit it during manual testing of the Antigravity runner in #167 and typed `--phase phase_03_review` by hand. No orchestrator or model chose the value. Cost was one retry — the `ValueError` lists the allowed values and self-corrected me. Full detail in the review thread.
- I took the directory name from `ROLE_ARTIFACTS` in source, not from disk; the quest directory did not exist yet. The tuple's first element sits next to the filenames, and nothing at that call site marks it as not-the-phase.
- Given the trigger was off-path, this is usability hardening found while building a third runtime, not a bug that hit a quest. Reasonable to close instead of merging — see the options in the review thread.

### DIFF
--- a/scripts/quest_runtime/artifacts.py
+++ b/scripts/quest_runtime/artifacts.py
@@ -113,7 +113,12 @@ def expected_artifacts_for_role(
             ) from exc
 
     normalized_phase = phase.strip().lower().replace("-", "_")
-    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent]
+    # The role's own artifact directory is accepted as a phase alias. It is the
+    # name a caller actually sees on disk (`.quest/<id>/phase_03_review/`), so
+    # passing it is the natural mistake; rejecting it cost real debugging time.
+    # Derived from ROLE_ARTIFACTS rather than hardcoded so renaming a phase
+    # directory cannot leave a stale alias behind.
+    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent] | {phase_dir}
     if normalized_phase not in allowed_phases:
         allowed = ", ".join(sorted(allowed_phases))
         raise ValueError(
--- a/scripts/quest_runtime/artifacts.py
+++ b/scripts/quest_runtime/artifacts.py
@@ -80,6 +80,16 @@ ROLE_PHASE_ALIASES: dict[str, frozenset[str]] = {
 }
 
 
+def _normalize_phase(value: str) -> str:
+    """Normalize a phase name for comparison.
+
+    Both sides of the phase check go through this, so an accepted alias and a
+    caller's input can never be compared under different rules.
+    """
+
+    return value.strip().lower().replace("-", "_")
+
+
 def default_quest_dir(workspace_root: str | Path, quest_id: str) -> Path:
     """Return the default repo-local quest directory for a run."""
 
@@ -112,13 +122,17 @@ def expected_artifacts_for_role(
                 f"Unsupported artifact subset {artifact_subset!r} for role {agent}"
             ) from exc
 
-    normalized_phase = phase.strip().lower().replace("-", "_")
+    normalized_phase = _normalize_phase(phase)
     # The role's own artifact directory is accepted as a phase alias. It is the
     # name a caller actually sees on disk (`.quest/<id>/phase_03_review/`), so
     # passing it is the natural mistake; rejecting it cost real debugging time.
     # Derived from ROLE_ARTIFACTS rather than hardcoded so renaming a phase
-    # directory cannot leave a stale alias behind.
-    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent] | {phase_dir}
+    # directory cannot leave a stale alias behind — and normalized through the
+    # same helper as the input, so a rename that introduced hyphens or case
+    # could not silently stop matching.
+    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent] | {
+        _normalize_phase(phase_dir)
+    }
     if normalized_phase not in allowed_phases:
         allowed = ", ".join(sorted(allowed_phases))
         raise ValueError(
--- a/scripts/quest_runtime/artifacts.py
+++ b/scripts/quest_runtime/artifacts.py
@@ -134,7 +134,11 @@ def expected_artifacts_for_role(
         _normalize_phase(phase_dir)
     }
     if normalized_phase not in allowed_phases:
-        allowed = ", ".join(sorted(allowed_phases))
+        # Hint from the RAW directory, not its normalized form. The point of
+        # listing it is to name something the caller can actually see in the
+        # tree; under a hyphenated rename the normalized value would name a
+        # directory that does not exist. Normalization is for matching only.
+        allowed = ", ".join(sorted(ROLE_PHASE_ALIASES[normalized_agent] | {phase_dir}))
         raise ValueError(
             f"Quest role {agent} is not valid for phase {phase!r}. Allowed: {allowed}"
         )
--- a/scripts/quest_runtime/artifacts.py
+++ b/scripts/quest_runtime/artifacts.py
@@ -123,9 +123,10 @@ def expected_artifacts_for_role(
             ) from exc
 
     normalized_phase = _normalize_phase(phase)
-    # The role's own artifact directory is accepted as a phase alias. It is the
-    # name a caller actually sees on disk (`.quest/<id>/phase_03_review/`), so
-    # passing it is the natural mistake; rejecting it cost real debugging time.
+    # The role's own artifact directory is accepted as a phase alias. The
+    # aliases below appear nowhere in the tree, while the directory name is
+    # visible both on disk (`.quest/<id>/phase_03_review/`) and in
+    # ROLE_ARTIFACTS right beside the filenames — so it is the natural guess.
     # Derived from ROLE_ARTIFACTS rather than hardcoded so renaming a phase
     # directory cannot leave a stale alias behind — and normalized through the
     # same helper as the input, so a rename that introduced hyphens or case
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -6,15 +6,29 @@ cd "$repo_root" || exit 1
 
 ./scripts/quest_validate-quest-config.sh || exit $?
 
-if python3 -m black --check .; then
+# Prefer the project virtualenv when present so venv-based workflows commit
+# without PATH juggling; otherwise use python3 from PATH.
+python3_bin="python3"
+if [ -x ".venv/bin/python3" ]; then
+  python3_bin=".venv/bin/python3"
+fi
+
+if ! "$python3_bin" -m black --version >/dev/null 2>&1; then
+  cat >&2 <<EOF
+black is not installed for ${python3_bin}.
+Install the pinned development dependencies:
+  ${python3_bin} -m pip install -e '.[dev]'
+EOF
+  exit 1
+fi
+
+if "$python3_bin" -m black --check .; then
   exit 0
 fi
 
-cat >&2 <<'EOF'
+cat >&2 <<EOF
 Python formatting check failed.
-Install the pinned development dependency:
-  python3 -m pip install -e '.[dev]'
 Format the repository, then retry the commit:
-  python3 -m black .
+  ${python3_bin} -m black .
 EOF
 exit 1
--- a/.github/scripts/security_ci_guard.py
+++ b/.github/scripts/security_ci_guard.py
@@ -597,6 +597,17 @@ _PIP_VCS_SCHEME_RE = re.compile(r"^(?:git|hg|svn|bzr)\+", re.IGNORECASE)
 _PIP_HTTP_URL_RE = re.compile(r"^https?://", re.IGNORECASE)
 
 
+def _strip_shell_quotes(token: str) -> str:
+    """Classify specs as pip receives them: the shell strips matching
+    surrounding quotes, so `'.[dev]'` reaches pip as `.[dev]`. Stripping
+    only converges on pip's view — it can reclassify a token as a local
+    path only when the shell would have handed pip that path anyway, so
+    quoting can never smuggle a remote spec into the local-path exemption."""
+    while len(token) >= 2 and token[0] == token[-1] and token[0] in {"'", '"'}:
+        token = token[1:-1]
+    return token
+
+
 def _is_pip_install_unpinned(line: str) -> bool:
     match = re.search(r"\b(?:python3?\s+-m\s+)?pip[0-9.]*\s+install\b(.*)$", line)
     if match is None:
@@ -609,7 +620,7 @@ def _is_pip_install_unpinned(line: str) -> bool:
     tokens = args_str.split()
     i = 0
     while i < len(tokens):
-        token = tokens[i]
+        token = _strip_shell_quotes(tokens[i])
         if token.startswith("-"):
             # `--requirement=req.txt` is a single token; nothing more to skip.
             if "=" in token:
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -41,16 +41,15 @@ jobs:
         with:
           python-version: '3.12'
 
-      - name: Install Black
-        run: python3 -m pip install black==26.3.1
+      - name: Create virtual environment
+        run: python3 -m venv .venv
+
+      - name: Install development dependencies
+        run: .venv/bin/python3 -m pip install -e '.[dev]'
 
       - name: Check Python formatting
-        run: python3 -m black --check .
-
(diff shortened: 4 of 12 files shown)


scripts/quest_runtime/artifacts.py
```diff
@@ -97,7 +97,12 @@ def expected_artifacts_for_role(
         raise ValueError(f"Unsupported quest role: {agent}") from exc
 
     normalized_phase = phase.strip().lower().replace("-", "_")
-    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent]
+    # The role's own artifact directory is accepted as a phase alias. It is the
+    # name a caller actually sees on disk (`.quest/<id>/phase_03_review/`), so
+    # passing it is the natural mistake; rejecting it cost real debugging time.
+    # Derived from ROLE_ARTIFACTS rather than hardcoded so renaming a phase
+    # directory cannot leave a stale alias behind.
+    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent] | {phase_dir}
```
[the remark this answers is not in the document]
zaphod72: Good catch — valid, and fixed in 042f78b.

You're right that it's latent today (every `phase_dir` in `ROLE_ARTIFACTS` is already lowercase with underscores), but it directly undercut the comment sitting above it. The whole reason for deriving the alias from `ROLE_ARTIFACTS` is that a rename can't leave a stale alias behind, and a hyphenated rename would have broken exactly the value the feature exists to accept:

```
ValueError: Quest role code-reviewer-b is not valid for phase 'phase-03-review'.
Allowed: code_review, phase-03-review, review, reviewing
```

Rejecting the name it simultaneously lists as allowed.

One deviation from the suggested diff: rather than duplicating `.strip().lower().replace("-", "_")` at the call site, both sides now go through a shared `_normalize_phase()` helper. Same result, but "both sides use the same rules" becomes structural rather than something a future edit has to remember to keep in sync — which is the property that just failed here.

Added a regression test that patches `ROLE_ARTIFACTS` with a hyphenated directory. Verified it fails without the source change and passes with it, so it's pinning the behaviour rather than just describing it. Suite 1087 → 1088.

scripts/quest_runtime/artifacts.py
```diff
@@ -96,8 +106,17 @@ def expected_artifacts_for_role(
...
+    normalized_phase = _normalize_phase(phase)
+    # The role's own artifact directory is accepted as a phase alias. It is the
+    # name a caller actually sees on disk (`.quest/<id>/phase_03_review/`), so
+    # passing it is the natural mistake; rejecting it cost real debugging time.
+    # Derived from ROLE_ARTIFACTS rather than hardcoded so renaming a phase
+    # directory cannot leave a stale alias behind — and normalized through the
+    # same helper as the input, so a rename that introduced hyphens or case
+    # could not silently stop matching.
+    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent] | {
+        _normalize_phase(phase_dir)
```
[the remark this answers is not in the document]
zaphod72: Right again, and it's a consequence of the previous fix rather than the original code — normalizing the derived alias closed the matching hole but fed the normalized value into the hint as well.

Fixed in d9b61bb by not letting matching and display share a value: membership is tested against the normalized set, the message is built from the raw `phase_dir`. Under a hyphenated rename the hint now reads `phase-03-review` — what's actually in the tree — rather than `phase_03_review`, which wouldn't exist.

Pinned by a test that patches `ROLE_ARTIFACTS` with a hyphenated directory and asserts the message contains `phase-03-review` and *not* `phase_03_review`, so the two can't quietly re-converge. Suite 1088 → 1089.

KjellKod:
rebased to have latest after some bugfixes regarding roles and state transitions

scripts/quest_runtime/artifacts.py
```diff
@@ -112,10 +122,23 @@ def expected_artifacts_for_role(
...
-    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent]
+    normalized_phase = _normalize_phase(phase)
+    # The role's own artifact directory is accepted as a phase alias. It is the
+    # name a caller actually sees on disk (`.quest/<id>/phase_03_review/`), so
+    # passing it is the natural mistake; rejecting it cost real debugging time.
+    # Derived from ROLE_ARTIFACTS rather than hardcoded so renaming a phase
+    # directory cannot leave a stale alias behind — and normalized through the
+    # same helper as the input, so a rename that introduced hyphens or case
+    # could not silently stop matching.
+    allowed_phases = ROLE_PHASE_ALIASES[normalized_agent] | {
```
KjellKod, line 134: Nice catch, and good hardening.

I'm genuinely curious how you hit it (@zaphod72). Across a couple hundred quests with Codex or Claude as orchestrator this never surfaced. My guess is Gemini in the mix? Different models interpret the same instructions differently.

The friction is real: the aliases the gate accepted (`code_review`, `plan_review`) appear nowhere on disk and nowhere in state.json, while the one name a caller can actually see (the phase directory) was the one value that failed. That's the perceived bug and this PR fixes it.

For anyone reviewing the quest repo and PRs, the widening is spelling-only, not permissions:

- Every role that lives in `phase_03_review` (code-reviewer-a/b, review-arbiter, fixer) now accepts `phase_03_review` alongside its existing aliases, and same for the `phase_01_plan` roles. That's the fix working as intended.
- What doesn't change: role permissions. Artifacts are still resolved by the `agent` argument, so `expected_artifacts_for_role(dir, "phase_03_review", "code-reviewer-a")` returns reviewer A's files and reviewer B gets B's. And a role only gets its own directory accepted, not everyone's: asking for a planner in `phase_03_review` still fails, and there's a test covering that.

Here's what's remaining, and what's blocking this from merging: we need to understand better how this triggered. Across hundreds of quests with Claude or Codex (various models) as orchestrator, and across many role/model combinations, we never hit it. I pulled the phase values from every historical `context_health.log` across four repos running the quest pipeline:

```
 879 plan_review    773 code_review    317 plan
 149 fix            128 build           50 review
  32 reviewing       24 building         6 fixing
```

Over 2,300 logged invocations, zero directory names. Likely two reasons: the workflow doc teaches the answer by example (`workflow.md` shows literal sample lines like `phase=plan_review | agent=plan-reviewer-a`, and the state.json names `building`/`reviewing`/`fixing`/`plan` are accepted aliases for most roles), and the ValueError lists the allowed values, so an orchestrator that guessed wrong once would self-correct on retry without anyone noticing.

From looking at the code, the only way to hit it is to invoke `quest_claude_runner.py` (or `expected_artifacts_for_role` directly) with `--phase phase_03_review`, by hand or from an agent poking at the runtime scripts outside a quest. That isn't the typical way a quest runs: in Codex, `$quest` (or `$quest <slug>` to resume) and in Claude `/quest` trigger the quest skill, and following `workflow.md` the orchestrator resolves the phase from the doc and state.json. The test suite can't raise it either. So I'd like to understand the path before merging: 
- _what were you running when it bit?_ 
- _If an agent hit this mid-quest, which model and what was it doing?_

To be clear, I'm genuinely glad this came up. Rough edges like this are exactly what a fresh setup with a different model surfaces, and this one is worth smoothing regardless. 

The one thing I want to _avoid_ is the runtime helpers gradually becoming an API for driving quests in ways the workflow _never intended_, because off-path invocation will surface a long tail of these and we'd be hardening each one after the fact to solve for potentially non-intended quest invocations. So my recommendation for anyone running with other models: always start from the quest skill entry point and let the orchestrator follow `workflow.md`. If something about that path didn't work for your setup, that's the more interesting bug and I'd love to hear about it.

KjellKod:
@zaphod72 to reduce PR friction, and since this PR has "Allow edits by maintainers" enabled, I resolved the merge conflicts with latest main and pushed directly to your branch. Additive merge commit only, your commits are untouched, no force push.

Context: items 2 and 3 from the PR description (the dead pytest option and the stale manifest entry) were also fixed in #165, which merged earlier today, along with the `.gitignore` entry. So after the merge this PR reduces to exactly fix 1: the phase-alias change in `artifacts.py` plus its tests. The only overlap to resolve was a duplicated `.gitignore` entry, dropped in the merge commit.

Verified on the merged branch: pytest 1197 passed, `tests/test-quest-runtime.sh` 72/72, `black --check` clean, manifest validator clean.

The question from my earlier comment still stands, I'm curious how you hit the phase-name rejection.

KjellKod (changes requested):
PR description needs to be updated.
Required sections are

- Summary (accepted: summary, overview)
- Changes (accepted: changes, what changed, key changes, files changed)

Use ## or ### headings. Matching is case-insensitive.

See .skills/pr-assistant/SKILL.md for the expected format.
Tip: use `/pr-assistant` to generate a compliant PR body.

scripts/quest_runtime/artifacts.py
zaphod72: **Short answer: I typed the phase name myself, in a manual test. No model chose it. No quest was running. Your data is right — this never bit a real quest.**

There is one decision for you at the end. Everything else is detail.

## What I ran

I was testing the new Antigravity runner for #167. It was a manual test of the dispatch contract, in a throwaway repo. It was not a quest.

```
python scripts/quest_antigravity_runner.py \
  --quest-dir .quest/e2e \
  --phase phase_03_review \
  --agent code-reviewer-b \
  --model gemini-3.6-flash-high \
  --add-dir /tmp/quest-e2e \
  ...
```

The command failed with the ValueError. The message listed the allowed values. I reran with `--phase code_review`. It worked. Total cost: one retry.

Note the script: `quest_antigravity_runner.py`, not `quest_claude_runner.py`. Same gate (`expected_artifacts_for_role`), reached from the sibling PR's code.

## It was not Gemini

Your guess was reasonable, but no. Gemini executed the role prompt. I typed the `--phase` value by hand. I never ran `/quest` during this work. So no model interpreted anything — a human read source code and guessed wrong.

## Where I got the wrong name

Small refinement to your framing. You said the caller sees the directory name on disk. I did not — the quest directory did not exist yet. I read `phase_03_review` from `ROLE_ARTIFACTS` in `artifacts.py`:

```python
"code-reviewer-b": (
    "phase_03_review",
    ("review_code-reviewer-b.md", ...),
),
```

The directory name sits right next to the filenames. Nothing at that site says it is not the phase. So the misleading name is in the source too — in the same dict this PR now derives the alias from.

## Two corrections to my own PR

You are deciding partly on my claims, so here are two that are wrong:

1. **The PR description.** It says all three fixes came from "setting up a clean checkout and running the suite." True for the pytest option and the manifest entry. False for this one — it came from #167 testing.
2. **The code comment.** It says the rejection "cost real debugging time." It cost one retry. Your error message self-corrected me — the same mechanism your 2,300-invocation data shows.

I will push both corrections on request, or fold them into whichever option you pick below.

## On your off-path concern

I agree with your rule: start from the quest skill entry point.

One distinction matters for #167. Calling the runner script directly is the documented interface — `workflow.md` tells the orchestrator to invoke it as a subprocess. The off-path part was narrower: no orchestrator resolved the phase. I supplied it by hand. That matches the gap I flagged in #167: the runner and probe are tested, but an orchestrator following the new prose is not. Your comment is a good argument for closing that gap rather than hardening the helpers under it.

## Your decision

This is usability hardening found while building a third runtime. It is not a bug that hit a quest. Three options, all fine with me:

1. **Merge as-is.** I correct the description and comment first.
2. **Split this fix out of #166.** The other two fixes land on their own.
3. **Drop this fix.** I close the thread. Your error message already self-corrects, which is what your data shows.

Tell me which and I will push it.

zaphod72: ^ That response is from Claude. My code changes typically start with a planning phase with many human reviews. Leans toward more smaller PRs. The implementation is done in a separate chat and a Claude `/review <PR#>` after each PR is pushed.

zaphod72:
Both resolved. Thanks for doing the merge — the reduction to fix 1 is right, and #165 covers the two I dropped.

**PR description** rewritten to the `pr-assistant` format: Why this matters / Summary / Changes / Validation / Notes. It now describes only the phase-alias change, notes that items 2 and 3 landed in #165, and records where this came from in the Notes section.

**Your question** — I answered it in the review thread at 19:30 UTC, which was after your comment, so you may not have seen it yet. Short version: I typed `--phase phase_03_review` by hand while manually testing the Antigravity runner for #167. No orchestrator, no model, no quest. One retry to correct. Your reconstruction was right on every point except the Gemini guess.

**One more correction pushed** (d23e0c3): the code comment said rejecting the directory name "cost real debugging time". It cost one retry. Replaced with the accurate reason the guess is natural — the aliases appear nowhere in the tree, while the directory name is visible both on disk and in `ROLE_ARTIFACTS` beside the filenames.

Verified on the merged branch: pytest 1197 passed, `tests/test-quest-runtime.sh` 72/72, `black --check` clean, manifest validator clean. Matches your numbers.

Also worth noting: #165 landed the venv-safe hook and single-sourced dev pins. I had both in this PR briefly, then reverted them when they broke six tests in `test_source_python_formatting.py` — I read the pinned contract as deliberate rather than something an outside contributor should change. Good to see it fixed properly at the source.

The merge-or-close question from the review thread still stands whenever you want to decide.

KjellKod:
Thanks for the careful work here, and especially for the honest follow-up. Your own answer to my question helped pick the resolution together with thinking about how we want to encourage usage of `quest` going forward., so I'm taking your option 3 and closing.

**The reasoning, in plain terms:**

The error never happened inside a real quest. It happened once, when a human typed `--phase phase_03_review` by hand while testing a new runner outside any quest setup. Across 2,300+ logged role invocations in four repos running the actual pipeline, no orchestrator ever passed a directory name. The workflow doc teaches the correct strings and the data says that works 100% of the time.

The failure already fixes itself if it would ever come up through agentic orchestration. The error message lists the accepted values. You hit it, read it, retried, done. Total cost one retry. A one-retry papercut in a hand-driven path doesn't justify permanent code.

The fix grew bigger than the problem,  a normalization helper, a raw-vs-normalized display split, and tests for a hyphenated directory rename that should never be needed as directory names are repo constants set by orchestration code, no human input ever flows through them. The new logic  defends against something that can't happen. 

For the record, items 2 and 3 from your original description were real and are already on main via #165, so this close loses nothing that mattered.

Genuinely glad you're poking at the edges, this kind of fresh-checkout friction report is valuable even when the answer is "won't fix". Looking forward to #167, where the actual gap you identified lives (an orchestrator following the new prose is untested, and that one I care about).

PR #167, opened 2026-08-08
Add Antigravity (Gemini) as a third Quest runtime

zaphod72:
## Summary

Adds the Antigravity CLI (`agy`) as a third runtime, so any of the nine canonical roles can be assigned a Gemini model in `.ai/allowlist.json`. Motivated directly by the OpenCode field notes' conclusion — *"two different reviewers beat two identical ones"* — a third independent family gives the reviewer slots somewhere else to disagree from.

Independent of #166; they touch different code and can land in either order.

## Changes

**One dispatch cell:**

| Orchestrator | Runtime | Entrypoint |
|---|---|---|
| Either | Antigravity | `python3 scripts/quest_antigravity_runner.py` |

Antigravity is **never an orchestrator**, only ever a dispatched runtime, so there's no per-orchestrator split, no MCP path, and no transport choice. That's why this is a lot smaller than the Claude runner it mirrors.

**Deliberately out of scope:** Antigravity as an orchestrator (would mean porting the quest skill to `agy`'s extension format and expanding the matrix from 4 cells to 9).

## Why a subprocess runner and not MCP

There is no official `agy` MCP server. But the setup guide already argues the CLI-bridge approach beats MCP for cross-model calls — filesystem scoping, per-invocation permission modes, handoff polling, fresh-process isolation. `agy --print` fits that shape exactly, so this follows the pattern you already chose and defended for Codex-led Claude roles rather than inventing a new one.

## Measured `agy` behaviour (docs were wrong on three counts)

Pinned to `agy 1.1.10`. These were measured, not read off the documentation:

| Question | Measured |
|---|---|
| Prompt via stdin? | **No.** Piping into `agy -p` runs an *empty* prompt. It must be the `-p` flag value, on argv — hence `MAX_PROMPT_ARGV_CHARS` rather than a bare `E2BIG` surfacing as an unexplained invocation error. |
| Where do errors go? | **stdout, inside the JSON envelope.** stderr came back empty for a rejected model. |
| Unknown `--model` exit code | **1** — which is what lets a rejection map onto the existing `model_rejected` kind. |
| `--mode plan` | **Not a write barrier** — see below. |

Because errors arrive on stdout, `classify_agy_result_kind()` is deliberately **not** a reuse of `classify_result_kind()`; that one inspects stderr and would have silently mis-classified every rejected model as `handoff_missing`.

## The containment finding

`--add-dir` is the **only** write boundary `agy` enforces:

- In `--mode plan`, `agy` still writes files inside `--add-dir` scope. Mode is posture, not permission.
- When a path is **outside** scope, `agy` doesn't refuse — it redirects the write to `~/.gemini/antigravity-cli/scratch/` and still reports `status: SUCCESS`, claiming it wrote the requested path.

That second one is a misdiagnosis trap: an unscoped dispatch surfaces much later as an unexplained `handoff_missing`, sending you to debug the model instead of the scoping. So `run_antigravity_role` **refuses to dispatch** with `invocation_error` when `add_dirs` is empty or doesn't cover the handoff directory, and `workflow.md` says a refusal is a scoping bug to fix rather than a model failure to retry.

**Not added:** a helper deriving `--add-dir` from `role_permissions`. `ideas/archive/2026-05-26-native-runtime-dispatch.md` records that reconciling slot ↔ `subagent_type` ↔ permission key is *"an enforcement concern, not a dispatch concern"*, so scoping stays orchestrator-supplied exactly as it is for the Claude bridge.

## Model IDs

Slugs pass to the CLI verbatim with no hardcoded allowlist, so a newly released model works the day it ships. (`gemini-3.5-pro-*` is due shortly and needs no code change — there's a test pinning that.) The bare `gemini` sentinel means "use the agy default", mirroring `claude`.

## Two things to disclose

- **Context isolation is weaker than the Claude path.** `agy` ingests workspace context on its own: a 595-byte probe prompt cost **75,883** input tokens inside a repo versus ~23k in an empty dir. The Context Retention Rule still bounds what the orchestrator retains, but it cannot bound what this runtime reads. Documented in `workflow.md`.
- **Preflight is cache-first here, unlike the Claude probes** where the cache is only a post-failure fallback. An `agy` probe is a real ~40s / ~76k-token model call, so re-proving a runtime that passed minutes ago would tax every quest start (measured 46s → 0.7s). TTL still bounds staleness and a cached entry is reusable only for the same probe model. Flagging it as a considered deviation, not an oversight — happy to revert to fallback-only if you'd rather the pattern stay uniform.

## Verification

- `pytest`: 1084 → **1109 passed** (+25), no regressions
- `black --check .`, `quest_validate-quest-config.sh`, `quest_validate-manifest.sh`: clean
- `test-quest-preflight.sh` 21/21, `test-quest-orchestration.sh` 35/35, `test-quest-runtime.sh` 50/50
- `test_quest_dispatch_guardrails.py` 18/18 — new prose kept in paragraphs separate from existing Codex-led text so its co-occurrence matchers don't false-positive; matcher untouched
- **Live end-to-end:** a real `code-reviewer-b` role prompt, built from `.skills/quest/agents/code-reviewer.md`, dispatched to `gemini-3.6-flash-high`. Clean pass in 71s — all three declared artifacts written, `context_health.log` tagged `runtime=antigravity`. It found both planted bugs with correct fixes plus one I hadn't planted, and returned findings in Quest's own richer schema (`finding_id`, `confidence`, `why_it_matters`, `evidence`, `needs_test`, `write_scope`) which the test prompt never specified — it read and obeyed the real role file's contract.

## Known gaps

- **An orchestrator following the new `workflow.md` prose is untested.** The runner, probe, preflight and dispatch selection are all exercised; a full `/quest` run driving them is not.
- **Codex-led → Antigravity is implemented but unexercised.** The mechanism is identical (a subprocess either way), so special-casing it would be more code for less value — but I have no Codex login to prove it.
- `.quest-checksums` intentionally untouched: on a clean checkout of `c67dd3e`, 64 of its 86 entries already mismatch, including files this branch never touches. Didn't want to add noise or misrepresent installer state.

## Open questions

1. **`opencode/gemini-*` collision** — provider-qualified OpenCode IDs would now classify as the Antigravity runtime. Right call, or should they stay with OpenCode?
2. **Ternary vs. a registry** — is a hardcoded third branch right, or would you rather generalise now there's more than two? Ternary is the smaller diff.
3. **Naming** — `antigravity` vs `gemini` as the runtime identifier. Went with `antigravity`: matches the binary, survives model renames.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<a href="https://cubic.dev/pr/KjellKod/quest/pull/167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


zaphod72:
Thanks — this was a genuinely useful pass. Nine of the ten findings were valid and are fixed in 0f35e7d. One had a factual error, noted at the end.

**Both P1s were real, and the first was a blocker.**

`write_default_from_allowlist` validates before writing but had no way to receive the probe result, so the new `antigravity_available=False` default rejected *every* Gemini-backed role — meaning a repo that configured one in its allowlist could never actually start a quest with it. That default was a deliberate fail-safe choice on my part and this is exactly the hole it left. Now a parameter, forwarded, with a test covering both the green-probe and red-probe paths.

The second P1 is the better catch of the two: a handoff alone was being treated as completion, so a role that wrote `handoff.json` but left its declared artifacts empty reported `handoff_json` and handed later stages empty files. Declared artifacts are now verified before a handoff is accepted, reporting the existing `artifact_missing` kind.

**The `add_dirs` generator finding deserves calling out** — it defeated the containment guard specifically. The check drained the iterator, so `build_agy_cmd` then dispatched with no `--add-dir` at all, having *just* validated one. Materialized once now, with a test that passes a real generator and asserts `--add-dir` reaches the command.

`select_role_runtime` was my own inconsistency: I widened the docstring to promise `antigravity` as a family name without handling it, so it fell through `runtime_for_model()` to Codex. `claude` and `codex` only worked by accident.

Also fixed: probe `artifact_missing` (now matching `run_bridge_probe`/`run_bg_probe` and the branch `quest_preflight.sh` already has), `text_fallback` as a distinct kind, UTF-8 byte counting for the argv guard, `json_quote_or_null` on the probe-path warning, finite-timeout validation, and `--probe`/`--orchestrator` as a usage error.

**One correction:** on the probe finding, the claim that `exit_code` becomes 0 isn't right. The line is `exit_code = 0 if probe_ok else returncode or 1` — with `probe_ok` False and `returncode` 0, `0 or 1` evaluates to `1`, so the exit code was already non-zero. The `result_kind` half of that finding was correct and is what I fixed.

Suite 1109 → 1115, all green.

zaphod72:
All three addressed in 8cdd02a.

**The P1 is the one that mattered, and it was a regression from my own previous fix.** Adding the artifact requirement caught the real case (a `complete` handoff with empty artifacts) but swept up two legitimate terminal outcomes with it. `needs_human` and `blocked` are both in `HANDOFF_STATUSES`, and a role that stopped to ask a question is *supposed* to have no finished artifacts — reporting `artifact_missing` would have routed it into retry/failure and buried the question it asked. Exempted, with a test covering `needs_human`, `blocked`, and the `complete`-with-no-artifacts case that must still fail.

One note on framing: the Claude runner doesn't actually apply a blanket artifact requirement when accepting a handoff — it uses artifacts for retry classification and in the probes — so "matching the Claude runner contract" isn't quite the right description. The finding is correct on its own merits regardless, which is why it's fixed.

**On the timeout finding — the crash doesn't reproduce.** I tested the largest finite double before changing anything:

```
int(1e308)                      -> fine, a 310-digit int
subprocess.run(['true'], timeout=1e308) -> accepted, no error
```

Python ints are arbitrary precision, so `int(timeout)` doesn't overflow for any finite float, and `subprocess.run` took the value. Anything that *would* overflow (`1e400`) parses to `inf` and was already rejected by the previous commit.

I've added the cap anyway, because a timeout of 10^308 seconds is a configuration error worth rejecting early rather than passing to `agy --print-timeout` — but as hygiene, not as a crash fix.

**P3 was a fair hit on my own sloppiness** — I'd pasted the same helper into both entrypoints. It now lives in `quest_runtime.antigravity_runner` and both CLIs import it, so the policy can't drift between the runner and the probe.

Suite 1115 → 1117, all gates green.

.skills/quest/delegation/workflow.md
```diff
@@ -70,10 +71,27 @@ If the preflight result was already cached by SKILL.md Step 2b, use the cached v
 6. If an older in-flight artifact has `transport_downgraded: true`, surface the preflight `warning` lines to the user once. New `auto` runs block instead of downgrading.
 7. Codex runtime roles use local Codex subagents. Do not probe, call, configure, or retry Codex MCP for Codex-led Codex roles.
 
+**Antigravity-backed roles (applies to both session types):**
```
KjellKod, line 74: **Must fix (prose-only).** This preflight block is correct, but its counterpart layer never learned about Gemini: `.skills/quest/SKILL.md` (override availability check, step 6) still says "Classify Claude-family model names as `claude` and `claude-*`; every other model name is Codex-backed." So at chooser time a `code-reviewer-b=gemini-3.6-flash-high` override is gated on Codex MCP availability: wrongly rejected if Codex is down, accepted with no agy validation otherwise. This dispatch-time block then acts as the only real gate.

Minimal fix is two sentences in SKILL.md: classify `gemini`/`gemini-*` as Antigravity-backed and require the antigravity probe result per this section. No new machinery — the dispatch gate here already exists as the second check.

scripts/quest_runtime/antigravity_runner.py
```diff
@@ -0,0 +1,624 @@
...
+        )
+
+    if resolved_artifact_paths:
+        try:
+            prepare_artifact_files(
+                resolved_artifact_paths,
+                quest_dir=resolved_quest_dir,
+                role=agent,
+            )
+        except OSError as exc:
```
KjellKod, line 372: **Must fix (1 line).** `prepare_artifact_files()` can raise `StateError` (malformed `plan_iteration`) and `PlanIterationError` (missing sealed snapshot at iteration > 1) for planner roles. The Claude runner catches both and returns a structured `invocation_error` envelope; this catches only `OSError`. Reproduced locally: a Gemini-assigned planner with a malformed `state.json` dies with an uncaught traceback instead of the JSON envelope orchestrators parse. Adding the two exception types to this clause restores parity.

.skills/quest/delegation/workflow.md
```diff
@@ -70,10 +71,27 @@ If the preflight result was already cached by SKILL.md Step 2b, use the cached v
...
+
+`--add-dir` is the **only** write boundary the Antigravity CLI enforces. Two measured behaviours make this non-negotiable:
+
+- **`--mode plan` is not a write barrier.** In plan mode `agy` still writes files that fall inside `--add-dir` scope. The runner selects plan mode for review and arbitration roles as a posture, not as a permission. Never describe it as read-only.
+- **An out-of-scope write is not refused, it is redirected.** When a target path falls outside `--add-dir`, `agy` writes to its own scratch directory (`~/.gemini/antigravity-cli/scratch/`) and still reports `status: SUCCESS`, claiming it wrote the requested path. The role then looks like it simply failed to produce a handoff.
+
+Therefore, every dispatch **must** pass `--add-dir` values that cover the quest directory and any other location the role's declared artifacts live in. `scripts/quest_antigravity_runner.py` refuses to dispatch and returns `invocation_error` when `--add-dir` is absent or does not cover the handoff directory — that refusal is a scoping bug in the dispatch, not a model failure, and must be fixed by correcting the `--add-dir` values rather than retried.
```
KjellKod: **Small adjustment (complexity/DRY).** Two things in this section:

1. The two "measured behaviours" bullets above are the fifth copy of the same explanation (module docstring, `READ_ONLY_ROLES` comment, `check_containment` docstring, quest_setup.md, here). Compress the rationale here to one or two sentences and let the module docstring in `antigravity_runner.py` stay the canonical version — the orchestrator only needs the instruction plus enough why to not describe plan mode as read-only.
2. "and any other location the role's declared artifacts live in" promises more than `check_containment` enforces (it validates the handoff directory only, which is the same directory as every current role's artifacts). Drop the clause so the prose matches the code — this is exactly the doc/code drift class #169 exists to catch.

KjellKod (changes requested):
Review by the local pair the CI gate would have been: fork PRs skip the codex-review job, so we ran its exact assembled prompt through gpt-5.6-sol locally, plus an independent Claude (Fable 5) review, then adversarially re-verified every finding before posting. Five findings survived the initial reviews; several were dropped or downgraded on a YAGNI pass. What's left is deliberately small.

**Verdict: request changes, scoped to the inline comments — roughly two sentences of prose fixes, one line of code, and one DRY compression.** The two must-fixes have first-run triggers, not hypothetical ones.

Notes that need a decision or a sentence, not code:

- **Open question 1 (`opencode/gemini-*`)**: recommend deciding, not coding. Today that ID reaches `agy --model` verbatim and is loudly rejected as `model_rejected`, self-explanatory. Keep the final-segment rule for consistency with `opencode/claude-*`; revisit only if someone actually runs Gemini through OpenCode.
- **Known limitation**: the runner has no `--artifact-subset`, so a Gemini-assigned arbiter's findings-only retry truncates and regenerates the valid verdict scratch too. Self-healing inefficiency; one sentence in workflow.md beats six lines of code until live testing says otherwise.

On the quality: this is disciplined work. The measured-not-documented CLI behavior, the containment refusal (the silent scratch-redirect trap is a genuinely nasty find), and the one-cell dispatch surface are all the right calls. Cache-first preflight deviation: keep it, the measurement justifies it. Ternary over registry: also keep.

Verified on this branch: pytest 1225 passed, preflight 21/21, orchestration 36/36, black clean, config validation clean. Live orchestrated testing (both orchestrator types) happens next on our side.

scripts/quest_runtime/antigravity_runner.py
```diff
@@ -0,0 +1,624 @@
...
+    # limit, so a multibyte prompt could pass a character count and still fail
+    # exec with a bare E2BIG that surfaces as an unexplained invocation error.
+    prompt_bytes = len(prompt.encode("utf-8"))
+    if prompt_bytes > MAX_PROMPT_ARGV_BYTES:
+        raise ValueError(
+            f"prompt is {prompt_bytes} UTF-8 bytes, over the "
+            f"{MAX_PROMPT_ARGV_BYTES}-byte argv limit for agy "
+            "(it has no --prompt-file equivalent)"
+        )
+    cmd = [
```
KjellKod, line 163: **Must fix (field-tested).** On a fresh install the runtime cannot write any artifact: agy 1.1.11, authenticated, default `settings.json` (no `permissions.allow` rules), headless `--print` auto-denies the `write_file` tool permission, so the preflight probe fails with:

```
jetski: no output produced — a tool required the "write_file" permission that headless mode
cannot prompt for, so it was auto-denied. Add an allow-rule under permissions.allow in
settings.json (e.g. write_file(<target>)). Alternatively, re-run with
--dangerously-skip-permissions to auto-approve all tools.
```

Verified both directions: quest's exact probe command shape fails; the identical command plus `--dangerously-skip-permissions` writes the file and returns SUCCESS in ~2s.

Proposed fix: add `--dangerously-skip-permissions` to the base `cmd` here, mirroring the Claude runner's `--permission-mode bypassPermissions` precedent. Containment is unchanged: per this PR's own measurements `--add-dir` is the only write boundary either way, so the flag doesn't weaken the model, it just removes the interactive-permission gate headless mode can't satisfy. (The alternative, `permissions.allow` rules in user-global settings.json, can't be scoped per-invocation to dynamic quest dirs, so the flag fits the dispatch model better.)

Question: you pinned agy 1.1.10 and your live e2e wrote artifacts. Did 1.1.10 not enforce headless permissions, or does your `settings.json` carry allow-rules? Worth knowing which, since it decides whether this is a 1.1.11 behavior change to note in the module docstring.

KjellKod: Confirmed fixed. Applied exactly the proposed change locally (`--dangerously-skip-permissions` in `build_agy_cmd`'s base cmd) and the preflight probe now passes live on the same fresh agy 1.1.11 setup:

```
"available": true, "probe_result_kind": "handoff_json"   # live probe, real artifact + handoff written
```

Cache behaviour verified too: the follow-up run returns `source: success_cache, cache_hit: true` with the TTL stamps, so the cache-first design works as described. The open question about 1.1.10 vs your local settings.json still stands for the docstring.

KjellKod:
Nice. 

I had to do one correction, but it worked really well. I don't have a Antigravity account and was running the free tier,and quest predictably enough fell back to defaults once Antigravity quota was reached. 

**quest 1 setup - codex as orchestrator**
```
  "models": {
    "planner": "gemini-3.1-pro-high",
    "plan-reviewer-a": "gemini-3.6-flash-high",
    "plan-reviewer-b": "gemini-3.6-flash-high",
    "arbiter": "gemini-3.6-flash-high",
    "builder": "gemini-3.1-pro-high",
    "code-reviewer-a": "gemini-3.6-flash-high",
    "code-reviewer-b": "gemini-3.6-flash-high",
    "review-arbiter": "gemini-3.6-flash-high",
    "fixer": "gemini-3.6-flash-high"
  },
  ```
  
  **quest 2 setup - claude as orchestrator**
  ```
    "models": {
    "planner": "gpt-5.6-sol",
    "plan-reviewer-a": "claude-opus-5",
    "plan-reviewer-b": "gemini-3.1-pro-high",
    "arbiter": "claude-opus-5",
    "builder": "gpt-5.6-sol",
    "code-reviewer-a": "claude-opus-5",
    "code-reviewer-b": "gemini-3.6-flash-high",
    "review-arbiter": "claude-opus-5",
    "fixer": "gemini"
  },
  ```

File: scripts/quest_runtime/antigravity_runner.py (hunk 4 above)
Comment (KjellKod) [reply scripts/quest_runtime/antigravity_runner.py:163]: Pushed the proposed fix to this branch as eabd92a (maintainer edit, additive commit, verified on a fresh agy 1.1.11 where the probe fails without it and passes with it). Suite: 1226 passed, black clean.

File: scripts/quest_runtime/antigravity_runner.py (hunk 2 above)
Comment (KjellKod) [reply scripts/quest_runtime/antigravity_runner.py:372]: Fixed in 3fc2343 (maintainer edit): `(PlanIterationError, StateError)` now return the structured `invocation_error` envelope, parity with `run_claude_role`. Pinned by a test that triggers `StateError` via a malformed `state.json`, no agy binary required.

File: .skills/quest/delegation/workflow.md (hunk 1 above)
Comment (KjellKod) [reply .skills/quest/delegation/workflow.md:74]: Fixed in dc0e728 (maintainer edit): SKILL.md's availability check now classifies `gemini`/`gemini-*` as Antigravity-backed and points at the antigravity probe. Field note: before this change a Claude-led session rejected gemini overrides with 'requires Codex MCP', exactly the failure predicted here.

**Comment by KjellKod** [comment]
Gemini instruction compliance vs codex: (codex orchestration)
```
  Handoff compliance:

  - Codex: 10/10, 100%
  - Antigravity: 6/10, 60%
  - Claude: 0/0, not used
  - Overall: 16/20, 80%
  ```
  
  not super impressive w/ 60% compliance but I ran it also with ALL the roles and multiple gemini models.

KjellKod:
Second model's read on the compliance numbers above (analyzed from the archived quest logs):

The 60% needs a denominator correction: all four Antigravity misses were the same failure, `RESOURCE_EXHAUSTED (429): Individual quota reached... Resets in 164h` (Starter weekly quota, burned by the plan phase plus a 15-minute pro-tier builder). Every dispatch that actually got model time wrote its handoff: 6/6. So the honest headline is **100% instruction compliance, 60% availability on a Starter quota**. The runtime followed the contract; the quota ran out.

Two things fall out of that:

1. `classify_agy_result_kind` has no quota branch, so the 429s surfaced as `handoff_json=unparsable`, pointing debugging at handoffs instead of at the quota message agy returned (which even includes the reset countdown). One more marker set (`"quota reached"` / `"resource_exhausted"`) mapping to a distinct result kind would let the orchestrator surface "quota exhausted, resets in Xh, reassign or wait" instead of a handoff mystery.
2. For `quest_setup.md`: an all-nine pro-tier quest exceeds a full Starter week. Flash-tier models and a role subset are the realistic Starter configuration.

Positives from the same run worth recording: the Codex-led → Antigravity cell executed for real (the PR's stated untested gap), the builder ran natively in a git worktree, the arbiter findings retry self-healed exactly as the known-limitation note predicted, and the Tier C fallback carried the quest to completion.

- Reviewed by Claude Fable 5, in collaboration with KjellKod

.skills/quest/SKILL.md
```diff
@@ -258,7 +258,7 @@ Before creating the quest folder, present the routing classification to the user
...
+   6. **Availability check.** Classify Claude-family model names as `claude` and `claude-*` (for example `claude-opus-5`); Gemini-family model names (`gemini`, `gemini-*`) as Antigravity-backed; every other model name is Codex-backed. Antigravity-backed models require `antigravity_available` from `scripts/quest_preflight.sh --probe antigravity` (see the workflow's Antigravity preflight section) in either session type. In Claude-led sessions, Claude-family models are available and Codex-backed models require Codex MCP availability from the Step 2b preflight result. In Codex-led sessions, Codex-backed models are available and Claude-family models require the top-level `available` boolean from the Step 2b preflight result, which represents Claude transport availability (background-agent or bridge — the `transport` field says which). If the relevant cache/result is missing or stale (older than the preflight TTL), rerun `scripts/quest_preflight.sh --orchestrator <self>` once and reuse the fresh result. Reject unavailable models with the preflight `warning` text and re-prompt the override submission.
```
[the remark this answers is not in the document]
KjellKod: Agreed, and verified against the preflight source rather than taking the finding on faith: `--probe` and `--orchestrator` are mutually exclusive modes, so the documented refresh sentence can never re-establish `antigravity_available`. This one is on us, the clause arrived in dc0e728. Fix incoming momentarily: the refresh instruction will name `--probe antigravity` for the Antigravity-backed case.

- Reviewed by Claude Fable 5, in collaboration with KjellKod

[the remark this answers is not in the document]
KjellKod: Fixed in f8bc923. The refresh sentence now names `--probe antigravity` for the Antigravity-backed case and states the mode exclusivity explicitly. Self-review notes: single occurrence of the refresh instruction in SKILL.md, the `.agents/skills/quest/SKILL.md` mirror is a stub with no availability paragraph (checked), workflow.md's own section was already correct as this finding noted. Guardrails 18/18, full suite 1226 passed.

- Reviewed by Claude Fable 5, in collaboration with KjellKod

.skills/quest/SKILL.md
```diff
@@ -258,7 +258,7 @@ Before creating the quest folder, present the routing classification to the user
...
...aude` and `claude-*` (for example `claude-opus-5`); Gemini-family model names (`gemini`, `gemini-*`) as Antigravity-backed; every other model name is Codex-backed. Antigravity-backed models require `antigravity_available` from `scripts/quest_preflight.sh --probe antigravity` (see the workflow's Antigravity preflight section) in either session type. In Claude-led sessions, Claude-family models are available and Codex-backed models require Codex MCP availability from the Step 2b preflight result. In Codex-led sessions, Codex-backed models are available and Claude-family models require the top-level `available` boolean from the Step 2b preflight result, which represents Claude transport availability (background-agent or bridge — the `transport` field says which). If the relevant cache/result is missing or stale (older than the preflight TTL), rerun `scripts/quest_preflight.sh --orchestrator <self>` once and reuse the fresh result — or `scripts/quest_preflight.sh --probe antigravity` when the stale result is the Antigravity one; the two are separate preflight modes and cannot be combined. Reject unavailable models with the preflight `warning` text and re-prompt the override submission.
```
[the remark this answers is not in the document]
KjellKod: Agreed and fixed in 78f3d6f. Verified the literal-reading path before accepting: the clause named only 'stale', so a missing cache (the most common case, a first-ever Gemini override) fell through to `--orchestrator <self>`. Rather than adding 'missing', the sentence now keys each command on the result it refreshes, so missing and stale are indistinguishable by construction.

- Reviewed by Claude Fable 5, in collaboration with KjellKod

.skills/quest/delegation/workflow.md
KjellKod: Done in e4a3742. The two measured-behaviour bullets are compressed to one sentence with the `antigravity_runner.py` module docstring named as the canonical copy (it was the fifth copy of the same explanation), and the 'any other location the role's declared artifacts live in' clause is dropped so the prose promises exactly what `check_containment` enforces. Guardrails 18/18, full suite 1226 passed.

- Reviewed by Claude Fable 5, in collaboration with KjellKod

KjellKod (approved):
All must-fixes from the earlier review are in (six maintainer-edit commits, field-validated on a real all-Gemini quest), CI is green, and every thread is resolved. Approving.

One welcome-but-not-blocking ask, here or as a follow-up, your choice: a quota branch in `classify_agy_result_kind` (`"quota reached"` / `"resource_exhausted"` markers) so a 429 surfaces as itself instead of `handoff_unparsable` — see the field report for the live case. The open questions from the PR description are answered in the threads: keep the final-segment rule for `opencode/gemini-*`, keep the ternary, keep the `antigravity` name.

- Reviewed by Claude Fable 5 + gpt-5.6-sol, in collaboration with KjellKod

PR #169, opened 2026-08-08
Pin the phase names workflow.md teaches to the ones the gate accepts

zaphod72:
## ⭐ Why this matters

`workflow.md` is not documentation *about* the dispatch contract — it is the copy an orchestrator learns the contract from. If it teaches a phase name the code rejects, every quest fails on the documented path, for every user. Nothing currently checks the two agree.

- **Catches doc/code drift** before an orchestrator hits it, in the required `test` check.
- **Test-only.** No production change, no new accepted inputs.

---

## Summary

Add a drift guard asserting that every phase name `workflow.md` teaches is a phase name `expected_artifacts_for_role()` accepts.

- `test_canonical_role_lockstep.py` already guards this family, but checks `ROLE_PHASE_ALIASES` **keys** (role slots) only. The **values** were unguarded.
- They agree today. Nothing stops them drifting apart with a green suite.

> [!NOTE]
> This is the follow-up suggested in #166: rather than making the runtime helpers tolerate more hand-typed input, pin the path the orchestrator is actually told to take.

## Changes

- **New drift guard** (`tests/unit/test_documented_phase_lockstep.py`):
  - Every `phase=` example in `workflow.md` is accepted by at least one canonical role.
  - Every worked `phase=X | agent=Y` example resolves for the role it names.
  - The extraction itself finds examples, so the other two cannot pass vacuously if the doc convention or the regex drifts.
  - Templated agent names in the doc (a truncated `agent=code-reviewer-` standing in for either slot) are filtered out rather than asserted on.

## Validation

- [ ] `python3 -m pytest tests/unit/test_documented_phase_lockstep.py` — 3 passed.
- [ ] `python3 -m pytest` — 1195 passed, no regressions.
- [ ] `python3 -m black --check .` — clean.
- [ ] Drift check: remove `"build"` from `ROLE_PHASE_ALIASES["builder"]`, and `test_every_documented_phase_is_accepted_by_some_role` fails on `phase=build`, which `workflow.md` still teaches. Restore to pass.

Watch for: the guard reads `workflow.md` at `.skills/quest/delegation/workflow.md`. If that file moves, the first test fails loudly rather than silently skipping.

## Notes

- **Deliberately not covered:** `REPLAN_PHASES` in `scripts/quest_runtime/state.py`. `plan_reviewed`, `presenting` and `presentation_complete` are in no alias set, and do not need to be — they are lifecycle states and never reach `expected_artifacts_for_role()`. Production has exactly one caller of that function (`quest_claude_runner.py`), taking `phase` from argv. Asserting those were dispatch phases would have pinned the wrong contract, so I checked before writing.
- Scope is phase *values*. Role *slots* stay guarded by `test_canonical_role_lockstep.py`, agent contract files by `test_runtime_agent_role_files_reference_canonical.py`. New file rather than an addition, to keep those scope notes accurate.
- Independent of #166 and #167 — touches no file either of them touches.

<a href="https://cubic.dev/pr/KjellKod/quest/pull/169?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tests/unit/test_documented_phase_lockstep.py
+++ b/tests/unit/test_documented_phase_lockstep.py
@@ -1,0 +1,111 @@
+"""Drift guard: every phase name ``workflow.md`` teaches must be a phase name
+``expected_artifacts_for_role`` accepts.
+
+Quest's control flow is markdown an orchestrator reads, so ``workflow.md`` is
+not documentation *about* the dispatch contract -- it is the copy the
+orchestrator learns the contract from. The accepted spellings live in
+``ROLE_PHASE_ALIASES``. Nothing currently connects the two, so an edit to
+either side can teach an orchestrator a value the code rejects, and the suite
+stays green.
+
+That failure mode is worse than an off-path mistake: the orchestrator follows
+the documented path faithfully and still fails, on every quest, for every user.
+These tests fail by name the moment the doc and the gate disagree.
+
+Scope: phase *values*. The role *slot* dimension is guarded by
+``test_canonical_role_lockstep.py``, and the agent-contract-file dimension by
+``test_runtime_agent_role_files_reference_canonical.py``.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+
+from quest_runtime.artifacts import expected_artifacts_for_role
+from quest_runtime.orchestration import CANONICAL_ROLES
+
+WORKFLOW_DOC = (
+    Path(__file__).resolve().parents[2]
+    / ".skills"
+    / "quest"
+    / "delegation"
+    / "workflow.md"
+)
+
+# Matches the context_health.log-style examples the doc uses to teach dispatch,
+# e.g. "phase=plan_review | agent=plan-reviewer-a".
+_PHASE_PATTERN = re.compile(r"phase=([a-z_0-9]+)")
+_PAIR_PATTERN = re.compile(r"phase=([a-z_0-9]+) \| agent=([a-z0-9-]+)")
+
+
+def _doc_text() -> str:
+    return WORKFLOW_DOC.read_text(encoding="utf-8")
+
+
+def _documented_phases() -> set[str]:
+    return set(_PHASE_PATTERN.findall(_doc_text()))
+
+
+def _documented_pairs() -> set[tuple[str, str]]:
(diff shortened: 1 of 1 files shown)


zaphod72:
Correct, and it undercut the whole point of the PR. Fixed in 6c51762.

The pattern only matched the context_health.log form (`phase=X | agent=Y`). It missed the "Split by role instance" block at workflow.md:1259-1268 entirely — the doc's own authoritative per-role list, and exactly the thing this PR claims to pin.

Your example is the right one to check against, so I injected it: swapped `review-arbiter` for `planner` under `phase=code_review` in `workflow.md`, rebuilt, and ran the suite.

Before the fix: passes. `phase=code_review` is still valid — three other roles accept it — and the wrong specific pairing was never extracted to be checked.

After the fix: fails, correctly —

```
workflow.md shows 'phase=code_review | agent=planner', but that pairing is rejected:
Quest role planner is not valid for phase 'code_review'. Allowed: plan
```

Reverted the injection and confirmed all three tests pass again. Pattern now accepts pipe or comma as the separator, so both doc forms are checked. Suite still 1195, black clean.

KjellKod (approved):
- Extraction is clean: 5 phases, 9 pairs covering all 9 canonical roles, and the one templated agent=code-reviewer- in the doc is correctly filtered, not asserted on. Zero false positives against today's doc (all 21 phase= occurrences).
- Mutation A (remove "build" from builder's aliases): caught, exactly as the PR's validation section claims.
- Mutation B (doc example with a typo phase): caught.
- Their claim about existing coverage is accurate: test_canonical_role_lockstep.py guards dict keys only, values were unguarded.
- Their scoping note is sound: lifecycle states (plan_reviewed, presenting...) never reach expected_artifacts_for_role, so excluding them pins the right contract. They checked the single production caller before writing, which is the discipline #166 lacked.
- Full suite on the branch: 1195 passed, matching their numbers.

KjellKod:
thanks @zaphod72

